PR #12, opened 2023-02-28
create initial README

bu-da:


bu-da:
decided to split it into different pull requests, that's why this pr is closed, but not merged.

PR #19, opened 2023-03-03
Feature bottom nav graph

nino-gerb:
I have to change the code to enable testing..

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/cook4me/ui/MainScreen.kt
+++ b/app/src/main/java/ch/epfl/sdp/cook4me/ui/MainScreen.kt
@@ -1,0 +1,69 @@
+package ch.epfl.sdp.cook4me.ui
+
+import android.annotation.SuppressLint
+import androidx.compose.foundation.layout.RowScope
+import androidx.compose.material.*
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.navigation.NavDestination
+import androidx.navigation.NavDestination.Companion.hierarchy
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.currentBackStackEntryAsState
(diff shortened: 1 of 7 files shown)


app/src/main/java/ch/epfl/sdp/cook4me/ui/MainScreen.kt
```diff
@@ -0,0 +1,70 @@
...
+fun BottomBar(navController: NavHostController){
+    val screens = listOf(
+        BottomBarScreen.Game,
+        BottomBarScreen.Home,
+        BottomBarScreen.Post,
+        BottomBarScreen.Profile,
+        BottomBarScreen.Search
+    )
+    val navBackStackEntry by navController.currentBackStackEntryAsState() //observes navBackStack entry
+    val currentDestination = navBackStackEntry?.destination //get the destination
```
pauromeu, line 35: Important: If navBackStackEntry is nullable, the case in which this variable is null should be handled.

app/src/main/java/ch/epfl/sdp/cook4me/ui/MainScreen.kt
```diff
@@ -0,0 +1,70 @@
...
+    Scaffold (
+        bottomBar = { BottomBar(navController = navController)}){
+        BottomNavGraph(navController = navController)
+    }
+}
+
+@Composable
+fun BottomBar(navController: NavHostController){
+    val screens = listOf(
+        BottomBarScreen.Game,
```
pauromeu, line 28: Proposition: The BottomBarScreen Game is defined in screens but not used in the BottomBar composable. If there is no plan to use it, it should be removed

pauromeu (approved):
Overall, the code looks well-written and follows best practices. Handling navigation is very important and this first approach will be useful when creating more complex structures. There are some suggestions for improvements.

PR #20, opened 2023-03-03
Bootcamp/web apis (database and tests not functional)

JohnTomasTaylor:


dayan9265:
Implementation looks good to me, small detail try to follow the naming convention for future branches and remove the commented code on the test that won't be used.

PR #38, opened 2023-03-06
use firebase and implement repository pattern

bu-da:
This pull request adds the following to the project:
* use the firestore api to store data
* introduce the [repository pattern](https://martinfowler.com/eaaCatalog/repository.html) to have an abstraction between data access layer and the services using it
* use the firebase emulator to test the data access, adapt ci config to run it also on cirrus
* extend the README with information about the local developer setup and the general software architecture


app/src/androidTest/java/ch/epfl/sdp/cook4me/repository/RecipeRepositoryTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        val newEntry1 = Recipe(name = "newEntry1")
+        val newEntry2 = Recipe(name = "newEntry2")
+        recipeRepository.add(newEntry1)
+        recipeRepository.add(newEntry2)
+        val allRecipes = recipeRepository.getAll()
+        assertThat(allRecipes.values, containsInAnyOrder(newEntry1, newEntry2))
+    }
+
+    @Test
+    fun updateExistingRecipe() = runTest {
```
dayan9265, line 60: maybe give bit more detailed name to tests (f.ex `updateExistingRecipeKeepsOnlyRecentRecipe()`)

app/build.gradle
```diff
@@ -74,22 +75,30 @@ configurations.forEach { config ->
...
-    testImplementation("org.hamcrest:hamcrest:2.2")
     implementation 'androidx.core:core-ktx:1.9.0'
     implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.5.1'
     implementation 'androidx.activity:activity-compose:1.6.1'
     implementation "androidx.compose.ui:ui:1.3.3"
     implementation "androidx.compose.ui:ui-tooling-preview:1.3.3"
     implementation 'androidx.compose.material:material:1.3.1'
+    implementation "androidx.navigation:navigation-compose:2.5.3"
+
+    implementation platform('com.google.firebase:firebase-bom:31.2.3')
```
dayan9265, line 86: Nitpick: maybe we could chose a way to represent strings (single quote or double quote) and be consistent, we could add this rule in the CI

dayan9265 (approved):
Except small detail (see comments), the code looks good to me

app/build.gradle
bu-da: Very good point 👍🏻  I created a new issue to check if we can enforce it via our tooling: https://github.com/cook4me/android/issues/43
I replaced all the "..." with '...' literals. It seems the following string is problematic otherwise: `"**/R$*.class"`

PR #45, opened 2023-03-14
Configure long method code smell

bu-da:
This pull request reconfigures codeclimate and detekt: 
* The long method check can be excluded from codeclimate, because it's already checked via detekt
* For composable functions it doesn't make sense to have a hard limit of 30 lines per function. Unfortunately, we can't exclude code from the detekt analysis by the `@composable` annotation. As a workaround we don't apply the check to the ui package.


Tachi-67 (approved):
LGTM

PR #47, opened 2023-03-14
Feature/auth

Tachi-67:
The user can now use signed in email and password to sign in with this feature.

The feature also creates notices for invalid email, blank password, successful or unsuccessful sign in, according to user inputs.

Took around 10 hours (implementation & tests)  :(

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/cook4me/ui/LoginScreen.kt
+++ b/app/src/main/java/ch/epfl/sdp/cook4me/ui/LoginScreen.kt
@@ -1,21 +1,34 @@
 package ch.epfl.sdp.cook4me.ui
 
-import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.Scaffold
 import androidx.compose.material.rememberScaffoldState
-import androidx.compose.runtime.*
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import ch.epfl.sdp.cook4me.R
-import ch.epfl.sdp.cook4me.util.*
+import ch.epfl.sdp.cook4me.util.BasicButton
+import ch.epfl.sdp.cook4me.util.BasicToolbar
+import ch.epfl.sdp.cook4me.util.EmailField
+import ch.epfl.sdp.cook4me.util.PasswordField
+import ch.epfl.sdp.cook4me.util.basicButton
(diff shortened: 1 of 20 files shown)


app/src/main/java/ch/epfl/sdp/cook4me/ui/LoginScreen.kt
```diff
@@ -0,0 +1,102 @@
...
+                horizontalAlignment = Alignment.CenterHorizontally
+            ) {
+                EmailField(email, { email = it }, Modifier.fieldModifier())
+                PasswordField(password, { password = it }, Modifier.fieldModifier())
+                BasicButton(
+                    stringResource(R.string.sign_in_screen_sign_in_button),
+                    Modifier.basicButton()
+                ) {
+                    scope.launch {
+                        if (!viewModel.isEmailValid(email)) {
```
JohnTomasTaylor, line 69: Nitpick: The error handling should probably be put into the viewModel, maybe by adding an invisible snackbar to the UI toggleable from the viewModel

app/src/main/java/ch/epfl/sdp/cook4me/util/ModifierExtension.kt
```diff
@@ -0,0 +1,12 @@
+package ch.epfl.sdp.cook4me.util
+
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+
```
JohnTomasTaylor, line 7: Not too sure about this pattern, feels less readable than having different button composables

app/src/main/java/ch/epfl/sdp/cook4me/ui/LoginScreen.kt
```diff
@@ -0,0 +1,102 @@
...
+                        } else if (viewModel.isPasswordBlank(password)) {
+                            scaffoldState.snackbarHostState.showSnackbar(context.getString(R.string.password_blank))
+                        } else {
+                            try {
+                                viewModel.onSignInClick(email, password)
+                                scaffoldState
+                                    .snackbarHostState
+                                    .showSnackbar(context.getString(R.string.sign_in_screen_sign_in_success))
+                            } catch (e: FirebaseAuthInvalidUserException) {
+                                scaffoldState
```
JohnTomasTaylor, line 82: Okay now this is just me being annoying, but I think it would be better just to show a "wrong username or password" message, so you can't try and guess valid usernames

JohnTomasTaylor (commented):
Thanks for the hard work! The code is super clean, the only slight nitpick I could find was having the firebase error handling in the loginScreen.

app/src/main/java/ch/epfl/sdp/cook4me/ui/LoginScreen.kt
Tachi-67: Thanks for the review, it is a nice idea to put the error handling inside the viewmodel.

However since things are working now I would suggest we make the refactoring another issue in out backlog and fix it sometime else.

app/src/main/java/ch/epfl/sdp/cook4me/util/ModifierExtension.kt
Tachi-67: Functions in this file are (for now) just for simplyfying boilerplates and, I don't feel the need to do refactorings. Maybe we can discuss it afterwards and decide if it is necessary to create an issue for this.

app/src/main/java/ch/epfl/sdp/cook4me/ui/LoginScreen.kt
Tachi-67: Firebase throws different exceptions for non-exist users (or deleted users) and wrong passwords, so I think it's better we handle them differently (since security is not our concern in this project, however user experience is)

JohnTomasTaylor (approved):
I'll add the refactoring to the backlog ^^

PR #68, opened 2023-03-21
Incorporate modification of pr review

dayan9265:
This pull requests fixes the problems that were not fixed in this previous merged pull request #65 (this pr was about the creation of a form to create event)


dayan9265:
I would appreciate some help with the use of `data class` since I am not used to it. I previously used a class `Event` to keep in memory the values associated with the creation of an event, but now with your suggestion to go with a data class I need to have my fields to be immutable, so here lies my question: 
- Since in my code (see `createEventScreen`), I modify field by field the `Event` object, how should I do that now? Create a copy of `Event` with one modification each time I want to modify a specific field, or is there an another solution?

bu-da (approved):
Looks very good overall, some minor issues in the comment. Thanks for taking the time! Don't forget to link the issue to the PR.

(Remark for the TA's/other readers of the PR: We had a look at the usage of data classes and other issues after yesterday's meeting. So they should be solved.)

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventCreationForm/CreateEventScreen.kt
```diff
@@ -1,88 +1,99 @@
...
+ * Component that shows the form to create an event
+ */
+@Composable
+fun CreateEventScreen() {
+    val event = remember {
+        mutableStateOf(Event())
+    }
+    val endMsg = remember { mutableStateOf("") }
+
+    fun updateDate(calendar: Calendar) = event.value.dateTime.set(
```
bu-da, line 40: Proposal: create a custom setter in the data class. This function should be grouped with the event class.

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventCreationForm/CreateEventScreen.kt
```diff
@@ -1,88 +1,99 @@
...
+        calendar.get(Calendar.YEAR), calendar.get(Calendar.MONTH),
+        calendar.get(Calendar.DAY_OF_MONTH)
+    )
+
+    fun updateTime(calendar: Calendar) {
+        event.value.dateTime.set(Calendar.HOUR_OF_DAY, calendar.get(Calendar.HOUR_OF_DAY))
```
bu-da, line 46: Proposal: same as above

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventCreationForm/CreateEventScreen.kt
```diff
@@ -1,88 +1,99 @@
...
+            text = stringResource(R.string.ask_event_number_participants), min = 2, max = 16,
+            onValueChange = { event.value = event.value.copy(maxParticipants = it) },
+            modifier = Modifier.fillMaxWidth()
+        )
+        ToggleButtonChoice(
+            question = stringResource(R.string.ask_event_visibility),
+            answerChecked = stringResource(R.string.event_visibility_everyone),
+            answerUnchecked = stringResource(R.string.event_visibility_subscriber_only),
+            onToggle = {
+                event.value = event.value.copy(isPrivate = it == "Subscriber only")
```
bu-da, line 75: Proposal: "Subscriber only" should be a ressource string too.

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventCreationForm/CreateEventScreen.kt
```diff
@@ -1,88 +1,99 @@
...
+            }
+        )
+        DatePickerComponent(
```
bu-da, line 78: Question: Why are these two composables called `...Component`? Can't you just call them `DatePicker` and `TimePicker`? The postfix Component doesn't add any info to the name.

dayan9265: The reason is that `DatePicker` was already a [class of android](https://developer.android.com/reference/android/widget/DatePicker), I don't know it is an issue or not (maybe since we won't be using this class it is not a problem)

bu-da: You could call it CustomDatePicker or sth like that

PR #74, opened 2023-03-24
Try to fix issues with google maps tests

bu-da:
We added a good amount of tests last week to our repo. Unfortunately since then our CI is failing randomly. This PR changes the following and tries to make the CI deterministic again:
* Fix race condition in `testCameraReportsMoving` test. Error case: 
   1. Test threads waits for `cameraPositionState.isMoving == true`
   2. UI thread sets `cameraPositionState.isMoving = true`
   3. UI thread sets `cameraPositionState.isMoving = false`
   4. Test thread asserts `cameraPositionState.isMoving = true`, which fails
* Increase overall timeouts. e.g. The timeout of 1 sec for initialising the map was too short and was causing issues in some branches.


app/build.gradle
```diff
@@ -86,9 +94,11 @@ dependencies {
     implementation 'androidx.compose.ui:ui-tooling-preview:1.3.3'
     implementation 'androidx.compose.material:material:1.3.1'
     implementation 'com.google.maps.android:maps-compose:2.11.0'
-    
     implementation 'com.google.android.gms:play-services-maps:18.1.0'
     implementation 'androidx.navigation:navigation-compose:2.5.3'
+    implementation "com.google.code.gson:gson:2.8.7"
+    implementation "androidx.navigation:navigation-compose:2.5.3"
+    implementation "io.coil-kt:coil-compose:2.2.2"
```
pauromeu, line 101: Proposition: We have a lot of comments from Cirrus CI about newer versions for the packages we are using. In a future refactoring it would be worth our time to check if we should increase the version.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/MapScreenTest.kt
```diff
@@ -73,42 +75,23 @@ class GoogleMapViewTests {
         initMap()
         assertEquals(CameraMoveStartedReason.NO_MOVEMENT_YET, cameraPositionState.cameraMoveStartedReason)
         zoom(shouldAnimate = true, zoomIn = true) {
-            composeTestRule.waitUntil(1000) {
-                cameraPositionState.isMoving
+            verify(ordering = Ordering.ORDERED, timeout = 60000L) {
```
pauromeu, line 78: Important: Time measures should be added as constants that can be placed in the top part of the file. Also, when it is not 100% clear, it would be nice to comment the units (s, ms, etc.)

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/MapScreenTest.kt
```diff
@@ -179,17 +152,19 @@ class GoogleMapViewTests {
...
-const val assertRoundingError: Double = 0.01
+const val ASSERT_ROUNDING_ERROR: Double = 0.01
 
-val hasValidApiKey: Boolean =
-    MAPS_API_KEY.isNotBlank()
-
-fun LatLng.assertEquals(other: LatLng) {
-    assertEquals(latitude, other.latitude, assertRoundingError)
-    assertEquals(longitude, other.longitude, assertRoundingError)
+private fun assertMoveHappened(cameraPositionState: CameraPositionState) {
```
pauromeu, line 157: Great! This repeating pattern can be better placed as a function.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/MapScreenTest.kt
```diff
@@ -179,17 +152,19 @@ class GoogleMapViewTests {
...
+    verify(ordering = Ordering.ORDERED, timeout = 60000L) {
```
pauromeu, line 158: Important: Same comment than before about the magic numbers in timeouts

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/MapScreenTest.kt
```diff
@@ -48,16 +48,18 @@ class GoogleMapViewTests {
                 }
             )
         }
-        val mapLoaded = countDownLatch.await(10, TimeUnit.SECONDS)
+        val mapLoaded = countDownLatch.await(60, TimeUnit.SECONDS)
```
pauromeu, line 51: Important: 60 seconds may be a value too high to deal with in the CI since this function (initMap) is used by all the testing maps. When map actually is failing to load, the tests are going to take a lot of time. We should find a tradeoff between making sure that it has time enough to pass and we are no stuck forever waiting for the timeout

pauromeu (approved):
Overall, very good implementation. I like your approach to try to get rid off this random errors we are getting from the CI. Let's hope it works and we get rid of this problem. The code quality is very good. I just added some little comments that may make things work a bit better. Thanks for your time.

app/build.gradle
bu-da: Good point, I'll create a new PR enforcing fixing the warning via CI. As far as I know they come from the android linter.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/MapScreenTest.kt
bu-da: I only partly agree with this, because it's not a magic number if we pass it as a named parameter. But in this case it makes sense, because we can add the unit to the name.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/MapScreenTest.kt
bu-da: Well I see your point, but I set it that high on purpose. The map should load all the time, otherwise there is something wrong in the tests or the setup. A lower value may get in the way. Let's see. We still can adapt it.

PR #77, opened 2023-03-26
Order files in packages, add common buttons

bu-da:
This PR is part of a bigger refactoring. It covers:
* order files/composables in packages
* share buttons between different composables


app/src/androidTest/java/ch/epfl/sdp/cook4me/application/TupperwareServiceTest.kt
```diff
@@ -45,16 +45,17 @@ class TupperwareServiceTest {
```
JohnTomasTaylor, line 45: This file should probably also be deleted

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/tupperwareform/TupperwareCreationScenarioTest.kt
```diff
@@ -31,66 +31,20 @@ import org.junit.Test
...
-        val mockService = MockTupperwareService(title, desc, tags, images)
-
         composeTestRule.setContent {
             CompositionLocalProvider(LocalActivityResultRegistryOwner provides registryOwner) {
                 // any composable inside this block will now use our mock ActivityResultRegistry
-                TupCreationScreenWithState(TupCreationViewModel(mockService))
+                CreateTupperwareScreenWithState(TupCreationViewModel())
             }
         }
         composeTestRule.onNodeWithTag("AddImage").performClick()
```
JohnTomasTaylor, line 50: Now that there isn't a MockService that checks if the data has been correctly submitted, we should probably add a line that asserts that the database has received the tupperware, or create a mockRepository

JohnTomasTaylor (approved):
LGTM

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/tupperwareform/TupperwareCreationScenarioTest.kt
bu-da: Very good point. I leave it like this until we have common way to handle state (coming soon). Then we probably should adapt the tests.

PR #78, opened 2023-03-27
Refactor/standardize state handling

bu-da:
This pr completes the refactoring for now. It adds an abstraction for input field form states, which takes care of the input validation. Due to time constraints, I could only refactor the login screen and the CreateTupperwareScreen.


app/src/main/java/ch/epfl/sdp/cook4me/ui/common/form/TextFieldState.kt
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ch.epfl.sdp.cook4me.ui.common.form
+
+import android.util.Patterns
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+
```
JohnTomasTaylor, line 22: Nice idea!

app/src/main/java/ch/epfl/sdp/cook4me/ui/common/form/TextFieldState.kt
```diff
@@ -0,0 +1,65 @@
...
+        get() = errorMsg
+
+    fun onFocusChange(focused: Boolean) {
+        isFocused = focused
+        if (focused) isFocusedDirty = true
+        if (!isFocused && isFocusedDirty) {
+            enableShowErrors()
+        }
+    }
+
```
JohnTomasTaylor, line 47: I think there's another way of doing this with InteractionSource.collectIsFocusedAsState() [like here](https://stackoverflow.com/questions/67339712/jetpack-compose-state-list-drawable-equivalent). That way we might not need use the onFocusedChanged modifier each time we add a textField

JohnTomasTaylor (approved):
Nice work! I see you ripped my viewModel to shreds :,(

PR #101, opened 2023-03-30
Feature/profile signup screen

nino-gerb:
I created a repository that stores/loads the profile data on/from the firestore database. Then I created a viewmodel that enables communication with the repository. The viewmodel loads the data from firestore when it's created. The different screens use this viewmodel to get the data form the firestore. 

In addition to that, I created three screens. One screen displays a recipe, the second screen provides the signup and the third is a duplicate form the profile editScreen() but used in the signUp process.


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/profile/EditProfileScreenTest.kt
```diff
@@ -68,23 +85,71 @@ class EditProfileScreenTest {
...
+
+        composeTestRule.onNodeWithStringId(R.string.btn_save).performClick()
     }
 
     @Test
-    fun clickingSaveButton() {
-        composeTestRule.setContent {
-            EditProfileScreen()
-        }
+    fun profile_test() {
```
Tachi-67, line 93: I don't understand what's happening with this test, for me it's just asserting if the input is functional.

Tachi-67 (changes requested):
Super impressive work! However do notice that something shouln't be pushed!!! I just marked them, please just remove them in a commit.

Testing is a bit weird to me, we could look at it together sometime.

PR #106, opened 2023-04-03
refactor tests for detailed event view with performScrollTo()

Tachi-67:


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/eventform/DetailedEventScreenTest.kt
```diff
@@ -48,69 +50,86 @@ class DetailedEventScreenTest {
...
+        composeTestRule.onNodeWithText(testEventParticipants).assertIsDisplayed()
     }
     @Test
     fun eventDateTimeIsDisplayed() {
         composeTestRule.setContent {
             DetailedEventScreen(event = testEvent)
         }
-        composeTestRule.onNodeWithStringId(R.string.event_time).assertExists()
-        composeTestRule.onNodeWithText(eventDate).assertExists()
+        composeTestRule.onNodeWithStringId(R.string.event_time).performScrollTo()
```
nino-gerb, line 130: You could create a helper function, so that you do not have to use everytime the same code. The test looks correct.

nino-gerb (approved):

PR #110, opened 2023-04-04
Feature/display loggedin user

Tachi-67:
This feature is built on top of feature/log_out, because I need to use the overviewViewModel.

With this feature the user can see on the overview screen the email used to log in.

We could probably in the future change it to the real name if necessary.

This took me ~1 hour to finish.

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/cook4me/ui/overview/OverviewScreen.kt
+++ b/app/src/main/java/ch/epfl/sdp/cook4me/ui/overview/OverviewScreen.kt
@@ -1,16 +1,22 @@
 package ch.epfl.sdp.cook4me.ui.overview
 
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
 import androidx.compose.material.Button
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/sdp/cook4me/ui/overview/OverviewScreen.kt
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+fun CurrentLoggedInEmailText(email: String?) {
+    var userEmail = email
+    if (userEmail == null) {
+        userEmail = "Empty Email"
```
dayan9265, line 105: I think this should be in the string resource

app/src/main/res/values/strings.xml
```diff
@@ -8,6 +8,7 @@
     <string name="navigate_to_add_event">Add event</string>
     <string name="navigate_to_add_signup">Sign up</string>
     <string name="navigate_to_postView">Post</string>
+    <string name="sign_out">Sign out</string>
```
dayan9265, line 11: It appears to be unused this string, I don't know if you want to later use it or it remains from a previous implementation?

dayan9265 (changes requested):
This code looks good to me 👍, just requested small change on the string resource, but otherwise well done :)

app/src/main/res/values/strings.xml
Tachi-67: It is a name from another branch, I merged it into this branch accidentally ^^', it's just fine, they will finally be used after I merged everything.

app/src/main/java/ch/epfl/sdp/cook4me/ui/overview/OverviewScreen.kt
Tachi-67: Good idea! Will do.

dayan9265 (approved):
Nice, thx for the small modif, you should be able to merge it now :)

PR #112, opened 2023-04-06
Add recipe screen navigation error fix

nino-gerb:
I added the screen to the navGraph so that the add does not crash anymore when calling the screen.


Tachi-67 (approved):
LGTM

PR #113, opened 2023-04-17
Feature/display actual event

Tachi-67:
This feature allows the user to display actual events from firestore to a screen.

What I did:
 - Extended `ObjectRepository` to allow for arbitrary queries from firestore, only need to specify the field to query, and the value you are looking for. Usage instructions? See comments in code.
 - Added a secondary constructor in Event.kt to convert a documentsnapshot data to an event. (for the sake of retriving data from firestore becase the original Event class could not be converted simply with toObject())
 - Extended EventFormService to allow for arbitrary queries.
 - Extended EventFormService to allow for arbitrary queries and getting the first event from query results.
 - Created the DetailedEventScreen and DetailedEventViewModel, to display the first event of the user.
 - Bug fix in CreateEventScreen: now the desciption input is normally functional.
 - Bug fix in IntegerSlider: now the slider is normally functional (before: the value will always change back to min)
 - Test for the feature, in the test I used the firebase emulator to create a local firestore.

The whole implementation took way more time (~7 hours) than expected (4 hours) because I had to deal with issues with the Event class and bugs in createEvent. The implementation of using firestore was also new to me, I had to struggle a bit.


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventScreenTest.kt
```diff
@@ -1,33 +1,67 @@
...
     private val calendar = Calendar.getInstance()
+    private lateinit var testEvent: Event
     private lateinit var eventDate: String
+
     @Before
     fun setUp() {
         context = InstrumentationRegistry.getInstrumentation().targetContext
-        calendar.set(Calendar.YEAR, 2023)
+        /*
+        * IMPORTANT:
```
pauromeu, line 45: Proposition: I understant that this try catch is already functional. This comment should be deleted to avoid confusion.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/signout/SignOutTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.tasks.await
+import kotlinx.coroutines.test.runTest
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@ExperimentalCoroutinesApi
```
pauromeu, line 29: Question: Why are we using Experimental apis? It is fine, I am just curious

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventScreenTest.kt
```diff
@@ -40,93 +74,83 @@ class DetailedEventScreenTest {
...
         }
-        composeTestRule.onNodeWithStringId(R.string.event_name).performScrollTo()
-        composeTestRule.onNodeWithStringId(R.string.event_name).assertIsDisplayed()
-        composeTestRule.onNodeWithText("test event name").performScrollTo()
-        composeTestRule.onNodeWithText("test event name").assertIsDisplayed()
     }
 
     @Test
-    fun eventDescriptionIsDisplayed() {
+    fun testUIIsCorrectlyDisplayed() {
```
pauromeu, line 110: Important: Names for tests should be a bit more precise

app/src/main/java/ch/epfl/sdp/cook4me/application/AccountService.kt
```diff
@@ -13,10 +13,10 @@ class AccountService(private val auth: FirebaseAuth = FirebaseAuth.getInstance()
 
     fun getCurrentUserEmail(): String? {
         val currentUser: FirebaseUser? = getCurrentUser()
-        if (currentUser == null) {
-            throw NullPointerException("current user is null!")
-        }
-        return currentUser.email
+        return currentUser?.email
```
pauromeu, line 16: Important: It seems like now we are not handling the case where current user is null.

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/ObjectRepository.kt
```diff
@@ -29,4 +30,37 @@ open class ObjectRepository(
...
+        val result = objectRepository.getWithGivenField<DataClass>(field, query)
+        return result.map { it.id to it.toObject(DataClass::class.java) }.toMap()
+      }
+     * This will return a map of id to DataClass object.
+     * If your dataclass sadly is not able to be serialized by firebase, add
+         a secondary constructor to your dataclass to construct it from a map.
+     * Want a demo? See EventFormService.kt and Event.kt
+    * */
+    suspend fun <A : Any> getWithGivenField(field: String, query: Any): List<DocumentSnapshot> {
+        val result = store.collection(objectPath).whereEqualTo(field, query).get().await()
```
pauromeu, line 63: Important: When expecting a resource from the DB, one should handle the case where the query fails and throws an error (p.e., "missing permissions" or "failed connection").

app/src/main/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventViewModel.kt
```diff
@@ -0,0 +1,30 @@
...
+    private val _firstEventState = mutableStateOf<Event>(Event())
+    val firstEventState
+        get() = _firstEventState
+
+    init {
+        viewModelScope.launch {
+            val userEmail = accountService.getCurrentUserEmail()
+            userEmail?.let {
+                val firstEvent = eventService.getFirstEventWithGivenField("id", userEmail)
+                firstEvent?.let {
```
pauromeu, line 24: Important: Cases where userEmail or firstEvent are null should be handled.

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventform/CreateEventScreen.kt
```diff
@@ -46,6 +48,10 @@ fun CreateEventScreen(
         event.value.dateTime.set(Calendar.MINUTE, calendar.get(Calendar.MINUTE))
     }
 
+    // for now I just set the id as the email of the current user for the sake of functionality
+    val userEmail = accountService.getCurrentUserEmail()
+    userEmail?.let { event.value = event.value.copy(id = userEmail) }
```
pauromeu, line 53: Important: When userEmail is null (I know it should not, but things may fail), the screen will crash because it is not handled

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventform/Event.kt
```diff
@@ -49,6 +50,28 @@ data class Event(
...
+            ?.let { calendarFromTime(it) }
+            ?: Calendar.getInstance(),
+        location = map["location"] as? String ?: "",
+        /*
+        * map["maxParticipants"]: get the value of the key "maxParticipants" in the map
+        * map["maxParticipants"] as? Long: cast the value to a Long, if it is not possible, return null
+        * (map["maxParticipants"] as? Long)?.toInt(): if the value is not null, cast it to an Int
+        * (map["maxParticipants"] as? Long)?.toInt() ?: 0 : provide a default value if the value is null (0)
+        * */
+        maxParticipants = (map["maxParticipants"] as? Long)?.toInt() ?: 0,
```
pauromeu, line 69: Proposition: maxParticipants is set to 0 in case of missing information. Maybe it would be more adequate to fix it to the arbitrary maximum value Int.MAX_VALUE

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventform/Event.kt
```diff
@@ -49,6 +50,28 @@ data class Event(
...
+            ?.toDate()
+            ?.let { calendarFromTime(it) }
+            ?: Calendar.getInstance(),
+        location = map["location"] as? String ?: "",
+        /*
+        * map["maxParticipants"]: get the value of the key "maxParticipants" in the map
+        * map["maxParticipants"] as? Long: cast the value to a Long, if it is not possible, return null
+        * (map["maxParticipants"] as? Long)?.toInt(): if the value is not null, cast it to an Int
+        * (map["maxParticipants"] as? Long)?.toInt() ?: 0 : provide a default value if the value is null (0)
+        * */
```
pauromeu, line 68: Important: Commented code like this should be removed from production

app/src/main/java/ch/epfl/sdp/cook4me/ui/overview/OverviewViewModel.kt
```diff
@@ -0,0 +1,20 @@
...
+    private val accountService: AccountService = AccountService()
+) : ViewModel() {
+    // navigationState = 0 -> no navigation
+    // navigationState = 1 -> signed out, navigate to login screen
+    private val _navigationState = mutableStateOf<Int>(0)
+    val navigationState
+        get() = _navigationState
+
+    fun onSignOutButtonClicked() {
+        accountService.signOut()
```
pauromeu, line 17: Important: You might want to consider adding error handling to this function, in case the sign-out process fails. For example, you could catch any exceptions thrown by the signOut method and handle them appropriately.

pauromeu (approved):
Thank you very much for your code! It seems like you worked hard during this sprint. The code appears to be well-structured and follows good programming practices. There is good separation of concerns between the different components, such as the ViewModel, services, and database access functions. I added some comments to the code, most of them referring to the need of handle errors from function that may fail.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventScreenTest.kt
Tachi-67: This is actually for future reference ^^ I'll add another comment saying it's already functional.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/signout/SignOutTest.kt
Tachi-67: I just used it from previous tests, iirc it prevents certain weird exceptions.

app/src/main/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventViewModel.kt
Tachi-67: If null, the firstevent will be not modified and stays being an empty event, ending up displaying nothing on the detailed event screen.

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventform/CreateEventScreen.kt
Tachi-67: actually it will not crash, since if the email is null, clauses in line 53 will not be executed and the event.value is not modified, the event id stays to be am empty string.

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventform/Event.kt
Tachi-67: This could have something to do with our business logic: if we find no max participants info about the event, should the event be seens as 'everybody is welcomed' or 'nobody for now, since there is something wrong with my configuration'? We could talk about this a bit.

app/src/main/java/ch/epfl/sdp/cook4me/ui/overview/OverviewViewModel.kt
Tachi-67: signout will not actually throw any exceptions. 

In my code's case, after signing out, the user is navigated to the login screen, and he/she will have to login again anyway.

In the case if there's a 'sign out failure' possibly because of network connection, the user is redirected to the login anyway, and has to login again. Logging in with the same credential will not cause any exceptions so everything is fine.

app/src/main/java/ch/epfl/sdp/cook4me/ui/eventform/Event.kt
Tachi-67: I don't really think it's a good idea to remove these comments as we might need to used similar syntax in the future, and these comments will become perfect references :)

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/ObjectRepository.kt
Tachi-67: You are right! Will do!

PR #114, opened 2023-04-18
Feature/user location

pauromeu:


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/map/MapPermissionWrapper.kt
```diff
@@ -0,0 +1,47 @@
+package ch.epfl.sdp.cook4me.ui.map
```
Tachi-67, line 1: Better to rename it as ....Test

Tachi-67 (approved):
Great work! Looks good to me.

PR #117, opened 2023-04-19
Create recipe feed

dayan9265:
This PR creates a recipe feed (similar to reddit) where the user can upvote / downvote recipe

Some functionalities are not complete in this PR:
 - The show `most recent recipe` button currently just display in the order which is inside the database (order of the id)
 - when clicking on a recipe it doesn't go to a recipe screen (not existing yet), but instead just display a dummy text ("This is a recipe")
 - It also currently does not remember if the user clicked on the upvote/downvote button when leaving the screen
 - detail there may be some concurrency problem if multiple people upvote at the same time, but I don't think this is currently important

So next sprint this feature could have some improvement 😉 

To have an idea on how it looks like:
image


app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/RecipeNoteRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+        store.collection(RECIPE_NOTE_PATH).whereEqualTo("id", id).get().await()
+            .first()?.getLong("note")?.toInt()
+    } catch (e: NoSuchElementException) {
+        println("No note for recipe $id (err: ${e.message})")
+        null
+    }
+
+    /**
+     * This method is used to add a note to a recipe.
+     * @param id the id of the recipe
```
nino-gerb, line 28: add @para note

app/src/main/java/ch/epfl/sdp/cook4me/application/RecipeFeedService.kt
```diff
@@ -0,0 +1,40 @@
...
+) {
+
+    /**
+     * Retrieves all the recipes and assigns them their notes (0 if they have none)
+     * @return a list of recipes with their id with their notes
+     */
+    suspend fun getRecipesWithNotes(): List<Pair<Pair<String, Recipe>, Int>> {
+        val recipes = recipeRepository.getAll()
+        val notes = recipeNoteRepository.retrieveAllRecipeNotes()
+        return recipes.map { Pair(Pair(it.key, it.value), notes[it.key] ?: 0) }
```
nino-gerb, line 22: look functional but not easy to read, use maybe a data class

app/src/androidTest/java/ch/epfl/sdp/cook4me/repository/RecipeNoteRepositoryTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun storeNewRecipeNotes() = runTest {
+        recipeNoteRepository.addRecipeNote("recipeId", 1)
+        recipeNoteRepository.addRecipeNote("recipeId2", 2)
+        val allRecipeNotes = recipeNoteRepository.retrieveAllRecipeNotes()
+        assertThat(allRecipeNotes.values, containsInAnyOrder(1, 2))
+        assertThat(allRecipeNotes.keys, containsInAnyOrder("recipeId", "recipeId2"))
+    }
+
+    @Test
+    fun updateRecipeNote() = runTest {
```
nino-gerb, line 57: name could be more explicit

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/recipeFeed/VoteIconTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun pressDownvoteDecreasesCounter() {
+        var counter = 0
+        composeTestRule.setContent {
+            VoteIcon(counterValue = 0, onChange = { counter += it })
+        }
+
+        composeTestRule.onNodeWithContentDescription("Downvote").performClick()
+        composeTestRule.onNodeWithText("-1").assertIsDisplayed()
+
+        assertThat(counter, `is`(-1))
```
nino-gerb, line 56: Maybe we should made that a recipe can not be less than zero. Because it look maybe a bit harsh

app/src/main/java/ch/epfl/sdp/cook4me/ui/recipeFeed/recipeFeed.kt
```diff
@@ -0,0 +1,144 @@
...
+ * Displays the recipe feed screen
+ * @param service the service to use to get the recipes/notes and update the notes
+ */
+@Preview(showBackground = true)
+@Composable
+fun RecipeFeed(service: RecipeFeedService = RecipeFeedService()) {
+    val isOrderedByTopRecipes = remember {
+        mutableStateOf(true)
+    }
+    val recipeList = remember {
```
nino-gerb, line 46: same here you could to a dataclass because your are using <Pair>Pair<String, Recipe> multiple

app/src/main/java/ch/epfl/sdp/cook4me/ui/recipeFeed/voteIcon.kt
```diff
@@ -0,0 +1,69 @@
...
+@Preview(showBackground = true)
+@Composable
+fun VoteIcon(counterValue: Int = 0, onChange: (Int) -> Unit = {}) {
+    val upvote = remember { mutableStateOf(false) }
+    val downvote = remember { mutableStateOf(false) }
+    val notPressedColor = Color.Black
+    val pressedColor = Color.Red
+    val localCounterValue = remember { mutableStateOf(counterValue) }
+
+    fun onVote(isUpVote: Boolean) {
```
nino-gerb, line 31: You could document this code. Add some comment to make this more clear.

nino-gerb (approved):
Overall your code looks good. The feature looks amazing! Some parts could be better documented, add maybe some comments. The PR looks really nice, i appreciate that you added an image.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/recipeFeed/VoteIconTest.kt
dayan9265: I agree that it can be a bit harsh, but at the same time we should be able to differentiate a bad recipe from a new recipe, so allowing negative notes allows that

dayan9265:
Good idea for the data class! It looks better and more clean now :), normally my code should be fine now

PR #120, opened 2023-04-20
Feature/create event from map screen

pauromeu:


app/src/main/java/ch/epfl/sdp/cook4me/ui/map/MapScreen.kt
```diff
@@ -88,6 +90,7 @@ fun GoogleMapView(
             .fillMaxWidth()
             .padding(vertical = 16.dp, horizontal = 16.dp)
     ) {
+        CreateNewItemButton(itemType = "Event", onClick = onCreateNewEventClick)
```
dayan9265, line 93: I think the "Event" should also be in the string resource

dayan9265 (approved):
Except the small detail of some strings that should be in the resource (you can directly merge after this small change), the code looks nice and it is a good idea to have to have created a generic add button 👍

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/MapScreen.kt
pauromeu: Yes, it should. Changed!

app/src/main/java/ch/epfl/sdp/cook4me/ui/common/button/CreateNewItemButton.kt
```diff
@@ -0,0 +1,48 @@
...
+        Row(
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            Icon(
+                imageVector = Icons.Default.Add,
+                contentDescription = "Create a new $itemType",
+                modifier = Modifier.size(24.dp)
+            )
+            Spacer(modifier = Modifier.width(8.dp))
+            Text(text = "Create a new $itemType")
```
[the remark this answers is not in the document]
pauromeu: Already changed. Thanks!

PR #134, opened 2023-04-21
enforce android lint on ci

bu-da:
With this PR we treat Android Studio lint warnings as errors on the CI. If we don't enforce this that way, we get a lot of technical debt over the time. This PR also initially fixes some issues complained by the linter or adds them to the baseline.


Tachi-67 (approved):
lgtm but I see a sad story for the upcoming sprints :(((

PR #138, opened 2023-04-21
show given event with event uid

Tachi-67:
I changed the previous detailedEventScreen to allow it to take the uid of an event on firestore as input, and displaying the actual event. The detailed event feature is now no longer dependent on the account service (which is good!)

I also added a circular waiting bar in the screen to indicate the data is still being retrieved in case of slow network connection. This is also the potential solution to "Correct profile info randomly showing" issue we found on Friday morning.

Usage:  DetailedEventScreen("someId")

Took me ~3 hours.

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
+++ b/app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
@@ -136,7 +136,8 @@ fun Cook4MeApp(
         composable(route = Screen.EditProfileScreen.name) { EditProfileScreen() }
         composable(route = Screen.CreateTupperwareScreen.name) { CreateTupperwareScreen() }
         composable(route = Screen.CreateEventScreen.name) { CreateEventScreen() }
-        composable(route = Screen.DetailedEventScreen.name) { DetailedEventScreen() }
+        // the uid of event is predefined on firestore. this is just for show.
+        composable(route = Screen.DetailedEventScreen.name) { DetailedEventScreen("IcxAvzg7RfckSxw9K5I0") }
         composable(route = Screen.SignUpScreen.name) {
             SignUpScreen(
                 onSuccessfullSignUp = { navController.navigate(Screen.SignUpUserInfos.name) },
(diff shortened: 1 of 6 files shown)


app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
```diff
@@ -136,7 +136,8 @@ fun Cook4MeApp(
         composable(route = Screen.EditProfileScreen.name) { EditProfileScreen() }
         composable(route = Screen.CreateTupperwareScreen.name) { CreateTupperwareScreen() }
         composable(route = Screen.CreateEventScreen.name) { CreateEventScreen() }
-        composable(route = Screen.DetailedEventScreen.name) { DetailedEventScreen() }
+        // the uid of event is predefined on firestore. this is just for show.
+        composable(route = Screen.DetailedEventScreen.name) { DetailedEventScreen("IcxAvzg7RfckSxw9K5I0") }
```
pauromeu, line 140: Important: It is fine to use hardcoded ids just for show but be should not forget to remove it

app/src/main/java/ch/epfl/sdp/cook4me/application/EventFormService.kt
```diff
@@ -43,6 +43,17 @@ class EventFormService(private val objectRepository: ObjectRepository = ObjectRe
...
     }
 
+    /*
+    * To get the event of given id.
+    * If nothing is found, null is returned
+    * */
+    suspend fun getEventWithId(id: String): Event? {
+        val result = objectRepository.getWithId<Event>(id)
+        // if(result == null) println("YESSSSSSSSSSSSSS")
+        // result?.let { println("NOOOOOOOOOOOO") }
```
pauromeu, line 53: Important: Please, don't push this debugging comments to production

pauromeu (approved):
Good job! The code looks well and this feature is going to be very useful to easily fetch events by id from the backend

app/src/main/java/ch/epfl/sdp/cook4me/application/EventFormService.kt
Tachi-67: Thanks, they are deleted in another commit.

app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
Tachi-67: Thanks for the notice, I’ll leave it here just for reference and remove it when we merge everything.

PR #139, opened 2023-04-22
Fix/profie load error

nino-gerb:
The data was not displayed correctly from the viewModel on the profile screen. 
 1. I changed init function of the Viewmodel that it relunches after 1 Second if the data is not received (This 1 second contition is maybe not so proper) 
 2.  I used remember for the viewModel in the two screens (profile screen and edit profile screen) . Therefore the lunch function gets called just one time. 
 3.  I added display now a loading bar on the profile screen when the data is loading. 
 4. I should still change the edithProfileScreen and do the refactoring better but this is part of the next issue
 5. I refactored the viewmodel to store a Profile instead of values of a profile 
 6. I refactored the editProfile screen to work with our state Fields
 7. I separated the test from editProfile screen and Profile  screen
 8. I added a callback function to the editProfile screen to navigate with the cancel button


app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
```diff
@@ -0,0 +1,92 @@
...
+    val formError: State<Boolean> = _formError
+    private val _profileState = mutableStateOf(Profile())
+    val profileState = _profileState
+
+    init {
+        viewModelScope.launch {
+            var profile = accountService.getCurrentUserWithEmail()?.let { repository.getById(it) }
+
+            // load the profile again if it is null
+            while (profile == null) {
```
Tachi-67, line 35: Is this really necessary since there is already an 'isLoad' variable in the screen?

Tachi-67 (approved):
Thanks for the hard work! The code is generally clean and nicely separated. 

There is one thing I don't quite understand though, in the profile viewmodel, is the delay really necessary since there's already an isLoad variable inside of the profile screen? Might be a good idea to try to remove it and see what happens, because I don't think hard-waiting for 1 second for the refresh is a neat idea.

app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
nino-gerb: I see your point that a delay with a magic number is not a good idea. What I wanted to achieve is that when the accountService function is loading again if it has a bug.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/profile/EditProfileScreenTest.kt
```diff
@@ -79,65 +75,107 @@ class EditProfileScreenTest {
 
     @After
     fun cleanUp() {
+        runBlocking {
```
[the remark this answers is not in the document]
nino-gerb: I'll have a look

PR #141, opened 2023-04-23
cache firebase executable

bu-da:


bu-da:
Aborted, because it doesn't give us any speed up on the CI

PR #152, opened 2023-04-26
Store tupperware data

bu-da:
This PR stores the tupperware data received via form in Firebase store and Firebase Cloud Storage (images). It adds the Cancel Button functionality.


storage.rules
```diff
@@ -0,0 +1,8 @@
+rules_version = '2';
+service firebase.storage {
+  match /b/{bucket}/o {
```
nino-gerb, line 3: Maybe we should consider security once

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareform/CreateTupperwareScreen.kt
```diff
@@ -110,13 +124,19 @@ private fun TupperwareForm(
...
     images: List<Uri>,
-    onSubmit: () -> Unit
+    onCancel: () -> Unit,
+    onSubmit: (String, String) -> Unit
 ) {
     val context = LocalContext.current
     val titleState = remember { RequiredTextFieldState(context.getString(R.string.TupCreateBlank)) }
-    val descriptionState = remember { RequiredTextFieldState(context.getString(R.string.TupCreateBlank)) }
+    val descriptionState =
+        remember { RequiredTextFieldState(context.getString(R.string.TupCreateBlank)) }
```
nino-gerb, line 133: Could you please explain to me why remember is needed?

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/TupperwareRepository.kt
```diff
@@ -1,18 +1,30 @@
...
-
-    suspend fun update(id: String, value: Tupperware) {
-        super.update(id, value)
+    suspend fun add(title: String, description: String, images: List<Uri>) {
+        val email = auth.currentUser?.email
+        checkNotNull(email)
+        val storageRef = storage.reference
+        images.forEach { path ->
+            val ref =
+                storageRef.child("/images/$email/$title/${UUID.randomUUID()}")
```
nino-gerb, line 25: Would it make more sense to use the userID as first repository?

nino-gerb (approved):
Overall it looks really nice. Thank you so much for the storage of the Images. I can really use it! It would be nice to have a closer look at the tupperwareform function together.

JohnTomasTaylor (changes requested):
The code is clean! Thanks for the hard work. 
There's just one thing bugging me about the tupperware creation test, I feel like it might be adding a tupperware to the real firebase database as it is using an unmocked recipeRepository. Could you maybe the check firebase console as I don't have access to it...

firebase.json
```diff
@@ -9,6 +9,12 @@
...
     },
-    "singleProjectMode": true
+    "singleProjectMode": true,
+    "storage": {
+      "port": 9199
+    }
+  },
+  "storage": {
+    "rules": "storage.rules"
   }
```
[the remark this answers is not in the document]
bu-da: This are the security rules for the firebase emulator and not the real cloud storage. But it would be probably a good idea to mirror the security rules we have there. At the moment, they are the same and they are the least restrictive they can be (everybody can write/read). We definitely should restrict this in the future so that only authenticated users can do this. I created an issue for this (we might need to adapt some tests and so it maybe takes some time): https://github.com/cook4me/android/issues/153

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareform/CreateTupperwareScreen.kt
```diff
@@ -98,7 +106,13 @@ fun CreateTupperwareScreen() {
...
             images,
-            onSubmit = {}
+            onCancel = onCancel,
+            onSubmit = { title, description ->
+                scope.launch {
+                    repository.add(title, description, images)
+                    onSuccessfulSubmit()
+                }
+            }
         )
```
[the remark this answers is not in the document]
bu-da: At the moment, I can only can think about over-engineered solutions, but maybe you can draft something?

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/tupperwareform/TupperwareCreationScenarioTest.kt
```diff
@@ -47,7 +47,7 @@ class TupperwareCreationScenarioTest {
         composeTestRule.setContent {
             CompositionLocalProvider(LocalActivityResultRegistryOwner provides registryOwner) {
                 // any composable inside this block will now use our mock ActivityResultRegistry
-                CreateTupperwareScreen()
+                CreateTupperwareScreen({}, {})
             }
```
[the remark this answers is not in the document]
bu-da: Very good point! Thanks for pointing it out! It seems not to, but due to time constraints I didn't investigate it any further. I adapted the test according to your suggestion.

storage.rules
bu-da: Very good point, I created an issue for this: https://github.com/cook4me/android/issues/153

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/TupperwareRepository.kt
bu-da: Maybe it does. We should sit together and discuss the general structure (but not as part of this pull request). Maybe sth like this?
> userId
   \ profile
   \ tupperware
      \title1
      \title2
      \title3

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareform/CreateTupperwareScreen.kt
```diff
@@ -98,7 +106,13 @@ fun CreateTupperwareScreen() {
             onClickTakePhoto = { onClickTakePhoto() },
             onClickImage = {},
             images,
-            onSubmit = {}
+            onCancel = onCancel,
+            onSubmit = { title, description ->
+                scope.launch {
+                    repository.add(title, description, images)
```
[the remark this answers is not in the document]
bu-da: we definitely can improve the robustness, but I would like to this later.

PR #157, opened 2023-04-28
Create a basic recipe screen and link it to the recipeFeed

dayan9265:
This create a recipe screen and when clicking on the recipe in the feed it displays the whole recipe (clicking again goes back to the default basic display)

The design is for the moment not the most beautiful, but the whole design of the app would be for an other PR ;)


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/recipe/RecipeScreenTest.kt
```diff
@@ -0,0 +1,51 @@
...
+        recipeSteps = listOf("Put the tomato on the dough", "Put the cheese on the tomato"),
+        difficulty = "Easy", cookingTime = "30min", servings = 4
+    )
+
+    @Test
+    fun titleCorrectlyDisplayed() {
+        composeTestRule.setContent {
+            RecipeScreen(dummyRecipe)
+        }
+        composeTestRule.onNodeWithText(dummyRecipe.name).assertExists()
```
Tachi-67, line 29: might be a good idea to use assertIsDisplayed() as assertExists() doesn't really check if anything is displayed on the screen.

Tachi-67 (changes requested):
Looks generally good to me, one thing to improve in the tests is it's a better idea to use assertIsDisplayed() rather than assertExists(), please consider changing it.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/recipe/RecipeScreenTest.kt
dayan9265: Yes true here it would be better. I had sometimes issue with `assertIsDisplayed()` if the screen is too big and doesn't completely fit in the emulator, then the `assertIsDisplayed()` returns false even though the component exists, but in this case it shouldn't be a problem, so I will change that 👍

Tachi-67: One thing that has always worked for me is to first call performscrollto() and then call the assertisdisplayed(), this will probably solve your issues when the screen is too big.

Tachi-67 (approved):
Thank you for the changes!!!!

PR #158, opened 2023-04-30
Feature/store recipe data

JohnTomasTaylor:
This PR finally adds recipe storage.


app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/TupperwareRepository.kt
```diff
@@ -19,12 +19,25 @@ class TupperwareRepository(
...
         images.forEach { path ->
             val ref =
-                storageRef.child("/images/$email/$title/${UUID.randomUUID()}")
+                storageRef.child("/images/$email/tupperwares/$tupperwareId/${UUID.randomUUID()}")
             ref.putFile(path).await()
         }
-        super.add(Tupperware(title, description, email))
+    }
+
+    override suspend fun delete(id: String) {
```
pauromeu, line 31: Important: Check this CodeClimate warning. Maybe it is worth it to create a function that can be reused so we gain a bit of maintainability

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/ObjectRepository.kt
```diff
@@ -11,10 +11,19 @@ open class ObjectRepository(
     private val objectPath: String = ""
 ) {
 
+    suspend fun <A : Any> addAndGetId(value: A): String {
```
pauromeu, line 14: Proposition: I am not sure of what happens when you receive an error from firestore. Shouldnt we handle this possible error when documentRef cannot be returned?

app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
```diff
@@ -175,7 +175,12 @@ fun Cook4MeApp(
                 onSignUpFailure = { navController.navigate(Screen.SignUpScreen.name) }
             )
         }
-        composable(route = Screen.CreateRecipeScreen.name) { CreateRecipeScreen(submitForm = {}) }
+        composable(route = Screen.CreateRecipeScreen.name) {
+            CreateRecipeScreen(
+                onSuccessfulSubmit = { navController.navigateUp() },
```
pauromeu, line 180: Question: Either when the create recipe screen is accepted or cancelled the user gets back to the previous screen? Will this work regardless of where is the initial location in the app?

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/RecipeRepository.kt
```diff
@@ -1,28 +1,45 @@
...
-        return result.map { it.id }.zip(result.toObjects(Recipe::class.java)).toMap()
-    }
-
-    suspend fun getById(id: String) =
-        store.collection(COLLECTION_PATH).document(id).get().await()
-            .toObject(Recipe::class.java)
-
-    suspend fun update(id: String, value: Recipe) {
-        store.collection(COLLECTION_PATH).document(id).set(value).await()
+    override suspend fun delete(id: String) {
```
pauromeu, line 33: Important: Make sure to handle the storage with Daniel so both of you are aware of what is going on with it.

pauromeu (approved):
Thank you for your great work! A lot of code is being added and it seems like it will increase the functionality of the app. Overall, very good work, I just have added a few comments.

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/ObjectRepository.kt
JohnTomasTaylor: Yes it is probably better to add a some error handling either in the repository part or one layer higher in the screens. I'll add that as an issue in our backlog.

app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
JohnTomasTaylor: Yes I thought it was more logical to bring them back to the screen in the app they came from, when successfully creating a recipe. I can't think of a case where it shouldn't work.

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/RecipeRepository.kt
JohnTomasTaylor: I'll do that

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/TupperwareRepository.kt
JohnTomasTaylor: I'll see if I find a smart way of refactoring that but for this sprint I'll concentrate on the offline part of the app first.

PR #159, opened 2023-04-30
Keep vote of user in the memory feed

dayan9265:
This code now stores the vote of the user using the address of the logged user


Tachi-67 (changes requested):
Thanks for the work!! The code looks nicely separated and well documented. One nitpicking though is to use less mockk for testing, but I don't see any fatal problems related.

Tachi-67 (approved):

PR #160, opened 2023-05-01
Feature/swipe on tupperware

bu-da:
This PR implements the actual swipe screen. The logic behind it is still missing (we're using hardcoded values at the moment), therefore it doesn't make sense to add tests for it.


app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareswipe/SwipeScreen.kt
```diff
@@ -0,0 +1,127 @@
...
+    fun swipe(direction: Direction) {
+        scope.launch {
+            val last = states
+                .reversed()
+                .firstOrNull {
+                    it.second.offset.value == Offset(x = 0f, y = 0f)
+                }?.second
+            last?.swipe(direction)
+        }
+    }
```
dayan9265, line 43: Maybe add some comments, just to make a bit clear how it works / what it does exactly

dayan9265 (approved):
Since this code is still in "beta" mode, for now it is fine. But later, it would require some cleaning, use of the string resource, testing... But for now it visually looks really cool, well done! Looking forward on the final result for this Tupperware swipe 😍 
And also good idea to have added the TODOs with the issues in the code to have an idea on future changes

Edit: we should also look on the reason of this failing test... (I created an issue for this), for now you can try to relaunch the CI and sometimes it passes

PR #162, opened 2023-05-02
Feature/Populate Map Firestore

pauromeu:


bu-da (approved):
I'll accept the PR now, because we need to changes for tomorrow. I'll do a complete review tomorrow morning, which should be considered next sprint.

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/ObjectRepository.kt
```diff
@@ -96,3 +98,21 @@ open class ObjectRepository(
             null
         }
 }
+
+open class ObjectCollectionRepository(
```
bu-da, line 102: Important: This should be part the `BaseRepository` (better name than `ObjectRepository`) or `EventRepository`

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/MapViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.lifecycle.viewModelScope
+import ch.epfl.sdp.cook4me.application.EventFormService
+import ch.epfl.sdp.cook4me.ui.eventform.Event
+import com.google.android.gms.maps.model.LatLng
+import kotlinx.coroutines.launch
+
+class MapViewModel(
+    eventService: EventFormService = EventFormService()
+) : ViewModel() {
+    private val _markers = mutableStateOf(emptyList<MarkerData>())
```
bu-da, line 15: Question: Do we really need the backing fields (fields with _ prefix)? Doesn't the following work? `val isLoading = mutableStateOf(true)`

bu-da (commented):
Looks good overall. The repository pattern is misunderstood, but this seems to be an issue of the whole team and needs to be addressed in a meeting.

PR #178, opened 2023-05-03
update firebase version in build.gradle because lint was not happy

Tachi-67:
emergency fix to main


bu-da (approved):

bu-da:
Did you check that the tests are still running through? Is the app still working? Version changes like this (we change the major version) can easily break functionality. We need to be careful.

Tachi-67:
> Did you check that the tests are still running through? Is the app still working? Version changes like this (we change the major version) can easily break functionality. We need to be careful. If you detect breaking functionality we can add an annotation with a comment to avoid linter complaints.

Both tried and yes it's working and the check is passing.

PR #180, opened 2023-05-03
Feature/store profile image on firebase

nino-gerb:
- I created a ProfileImageRepository, which stores one images on the firebase storage. There is a function to add a single image, to delete all images from a specific user at the profileimages location, and retrieve a image. 
- I separated the profileImage from the profiledata
- I add the profile image in the profileEdit screen and in the profileAddInfosScreen
- I retrieve the initcall of the profileViewModel
- In addition to the image stuff I fetched the userdata from the viewmodel to the profileedit screen.


app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
```diff
@@ -19,38 +21,69 @@ class ProfileViewModel(
...
+                        _profileState.value.email = it.email
+                        _profileState.value.name = it.name
+                        _profileState.value.allergies = it.allergies
+                        _profileState.value.bio = it.bio
+                        _profileState.value.favoriteDish = it.favoriteDish
+                    }
+                }
+            } catch (e: NoSuchElementException) {
+                if (e.message == "Collection is empty.") {
+                    onFailure()
```
Tachi-67, line 62: what happens here when the onfailure is called? not so sure about the logic here.

Please create an issue to really do something here (like navigating to somewhere, or popping out a message)?

Tachi-67 (approved):
Thank you for the hard work!!! This is really a biiig pr and I can tell the workload behind it^^

The code is generally really nice. No real serious issues.

However there is one small thing: In the profile view model, how the error is handled seems rather unclear. It is fine now because it doesn't make sense if a register user doesn't have any profile online, but it might be cooler if something is done there, like popping up some message.

Also, I'd recommend adding more comments explaining what's happending in the code, it's been quite a pain to go through the code without any explaination.

PR #192, opened 2023-05-03
Avoid raceCondition in AddProfileInfoTest

nino-gerb:
There is a racecondtion by the testTextFieldsInput test in AddProfileInfoScreenTest. I try to fix it.


app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
```diff
@@ -77,18 +78,20 @@ class ProfileViewModel(
...
-                        it, // Email as id
-                        profile.value.name,
-                        profile.value.allergies,
-                        profile.value.bio,
-                        profile.value.favoriteDish,
-                        profile.value.userImage
-                    )
-                    onSuccessListener()
-                    isLoading.value = false
+                runBlocking {
```
Tachi-67, line 81: Everything is fine. Just for your information that anything might come wrong again, I was also debugging with this (but in the detailed event viewmodel), what actually solved the problem was adding a delay(1000) just before the query ;)

Tachi-67 (approved):
Good idea of adding the runBlocking and waitforIdle(). Thanks for the quick fix ;)

app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
nino-gerb: But isn't it like that if i add run blocking the isLoading.value is only updated when the service.submitFrom executed completely?

PR #194, opened 2023-05-03
Feature/start chat from detailed event screen

Tachi-67:
This is an integration of chat and the detailed event screen.

What I did:
- Create a button in the detailed event screen to allow the user to create and start a chat with the event owner.
- Sometimes the detailed event screen keeps loading, I tried to fix the bug, now it works fine.
- Changed the loading signal in the detailed event screen (before was a weird dot) to a better screen (used in the chat as well)
- Stored the chat api key in a safe way (in local.properties)

The coverage of detailed event screen is 88% (which is exactly the same as what we have on the main), don't understand why the low coverage it shows here. This pr shouldn't be changing any coverage we have on main.


app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
```diff
@@ -160,7 +160,7 @@ fun Cook4MeApp(
         }
         composable(route = Screen.CreateEventScreen.name) { CreateEventScreen() }
         // the uid of event is predefined on firestore. this is just for show.
-        composable(route = Screen.DetailedEventScreen.name) { DetailedEventScreen("IcxAvzg7RfckSxw9K5I0") }
+        composable(route = Screen.DetailedEventScreen.name) { DetailedEventScreen("pSkhty73UrT4f55lIOov") }
```
nino-gerb, line 163: Yes i think it's already an improvement. But it still should be in a file apart I guess?

app/src/main/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventScreen.kt
```diff
@@ -27,11 +33,12 @@ fun DetailedEventScreen(
     eventId: String,
     detailedEventViewModel: DetailedEventViewModel = DetailedEventViewModel(eventId),
 ) {
+    val context = LocalContext.current
     val event = detailedEventViewModel.eventState.value
     val isLoading = detailedEventViewModel.isLoading.value
     Box {
         if (isLoading) {
-            CircularProgressIndicator(modifier = Modifier.align(Alignment.Center))
+            LoadingScreen()
```
nino-gerb, line 41: Your screen is way more beautiful than the circularProgressIndicator. I like that thanks for your hard work.

app/src/main/java/ch/epfl/sdp/cook4me/ui/detailedevent/DetailedEventScreen.kt
```diff
@@ -70,6 +77,22 @@ fun DetailedEventScreen(
 
                 SectionWithTitle(title = stringResource(R.string.event_time), content = event.eventDate)
                 Divider(color = MaterialTheme.colors.secondary, thickness = 1.dp)
+
+                Button(
+                    onClick = {
+                        createChatWithEmail(
+                            targetEmail = event.id,
+                            client = provideChatClient(
+                                apiKey = BuildConfig.CHAT_API_KEY,
```
nino-gerb, line 86: looks better

nino-gerb (approved):

app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
Tachi-67: yes, this is just for show. After we link event and maps this will be deleted

PR #202, opened 2023-05-04
remove chat from overview & add chat in buttom bar

Tachi-67:


JohnTomasTaylor (approved):
LGTM this will have to be merged with #195

PR #203, opened 2023-05-04
Fix/get rid of overview in chat nav

Tachi-67:
Previous code couldn't function: when I open chat, go back, and hit chat button again, the app doesn't navigate to the chat screen any more, but to the overview screen (weird).
I changed to navigate back to the recipe feed screen, seems to have solved the problem.


Tachi-67:
could be a better idea to popback? will try a bit.

Tachi-67:
> could be a better idea to popback? will try a bit.

Not really, aborted.

pauromeu (approved):
Quick fix! Looks good.

PR #232, opened 2023-05-10
TupperwareInfos 

nino-gerb:
Adding the TupperwareInfos screen
This screen has a link to the profile of the user who added the screen. Therefore I had to extend the ProfileScreen to show the profile on any user. 

Screenshot 2023-05-10 171431


app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareswipe/TupperwareInfos.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import ch.epfl.sdp.cook4me.persistence.model.Tupperware
+
+@Composable
+fun TupperwareInfos(
+    modifier: Modifier = Modifier,
+    tupperware: Tupperware = Tupperware("Title", "Description", "User"),
```
bu-da, line 23: Question: Why do you use a default parameter for the tupperware? To simplify testing? There should always be a tupperware you pass to this composable, if no tupperware exists then no infos are needed.

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareswipe/TupperwareInfos.kt
```diff
@@ -0,0 +1,56 @@
...
+            modifier = modifier
+                .fillMaxWidth()
+                .padding(16.dp)
+        ) {
+            Text(text = tupperware.title, style = MaterialTheme.typography.h6)
+            Spacer(modifier = modifier.height(8.dp))
+            Text(text = tupperware.description, style = MaterialTheme.typography.body1)
+            Spacer(modifier = modifier.height(16.dp))
+            Text(
+                text = "View Profile",
```
bu-da, line 42: Important: Move the string to the strings.xml file

app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
```diff
@@ -34,15 +37,20 @@ class ProfileViewModel(
...
+            } else {
+                // otherwise, use the currents user email
                 accountService.getCurrentUserWithEmail()?.let { repository.getById(it) }
+            }
 
             // load the profile again if it is null
             while (profile == null) {
-                @Suppress("MagicNumber")
-                delay(2000) // Wait for 1 second before retrying
+                delay(DELAY_MILLIS) // Wait for DELAY_MILLIS before retrying
```
bu-da, line 53: Question: Why do you need to retry? We don't do these anywhere else. If you really need this you should stop at the some point otherwise the user may end up in a endless loop

app/src/main/java/ch/epfl/sdp/cook4me/ui/profile/ProfileViewModel.kt
```diff
@@ -66,7 +74,7 @@ class ProfileViewModel(
             }
 
             try {
-                _profileImage.value = profileImageRepository.get()
+                _profileImage.value = profileImageRepository.get(id)
```
bu-da, line 77: Important: The application of the repository pattern is incorrect. There should exist only one repository regarding the profile called ProfileRepository or maybe even UserRepository. If you want to get the profile image you do it in the getById function or you create a new one called getImageById(), to specify that we only return a subset of the data.

Feel free to reach out to me, if the idea of the repository is still unclear.

bu-da (approved):
The changes look ok to me. The code of the profile definitely needs a refactoring.

PR #234, opened 2023-05-10
add instruction of chat api key to readme

Tachi-67:


nino-gerb (approved):
Thanks for adding it :)

PR #237, opened 2023-05-12
Fix/delete profiles bug

nino-gerb:
I used some function that deleted all profiles in the tests. Now I remove just the profiles i created. But I think there is still an issue with the way I connect to the emulator:

Sometimes the emulator does not connect and I think that then firebase connect to the real database

### DIFF
--- a/app/src/androidTest/java/ch/epfl/sdp/cook4me/repository/ProfileRepositoryTest.kt
+++ b/app/src/androidTest/java/ch/epfl/sdp/cook4me/repository/ProfileRepositoryTest.kt
@@ -46,6 +46,7 @@ class ProfileRepositoryTest {
             favoriteDish = "turkey",
         )
         runBlocking {
+            // get the profile from the database
             profileRepository.add(newEntry1)
             profileRepository.add(newEntry2)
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/ch/epfl/sdp/cook4me/repository/ProfileRepositoryTest.kt
```diff
@@ -33,16 +29,6 @@ class ProfileRepositoryTest {
         profileRepository = ProfileRepository(store)
     }
 
-    @After
-    fun cleanUp() {
-        runBlocking {
-            val querySnapshot = store.collection(COLLECTION_PATH).get().await()
-            for (documentSnapshot in querySnapshot.documents) {
-                store.collection(COLLECTION_PATH).document(documentSnapshot.id).delete().await()
```
pauromeu, line 41: Important: When awaiting for a firestore function it is useful to handle the errors so if it fails it is easier to catch.

pauromeu (approved):
Hey Nino! Your code looks good and clean. I hope this will solve the issue caused by this annoying bug. I just added a pair of comments regarding the handling of possible errors.

PR #238, opened 2023-05-12
Feature/create voting screen

nino-gerb:
Screenshot 2023-05-12 212457

I created a voting screen that takes a challenge and returns add  the votes from a participant to a new challenge which is return in a callback function.


app/src/main/java/ch/epfl/sdp/cook4me/ui/challenge/VotingScreen.kt
```diff
@@ -0,0 +1,174 @@
...
+            Text(
+                text = title,
+                style = MaterialTheme.typography.h6.copy(fontWeight = FontWeight.Bold),
+            )
+        }
+    }
+}
+
+@Composable
+private fun toolbarColor(darkTheme: Boolean = isSystemInDarkTheme()): Color =
```
pauromeu, line 173: Great! It is nice that you take all these details into account :)

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/challenge/VoteScreenTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        composeTestRule.onNodeWithContentDescription("participant1 Star 1").assertIsDisplayed()
+        composeTestRule.onNodeWithContentDescription("participant1 Star 2").assertIsDisplayed()
+        composeTestRule.onNodeWithContentDescription("participant1 Star 3").assertIsDisplayed()
+        composeTestRule.onNodeWithContentDescription("participant1 Star 4").assertIsDisplayed()
+
+        // Assert that the last 2 stars are empty
+        composeTestRule.onNodeWithContentDescription("participant1 Empty star 5")
+            .assertIsDisplayed()
+
+        composeTestRule.onNodeWithText("Vote").assertIsDisplayed()
```
pauromeu, line 129: Proposition: The test looks good and it passes but it is a bit difficult to read. There are a lot of lines doing almost the same. Maybe the use of function could make it more easy to maintain?

pauromeu (approved):
Hello Nino! Thank you for your contribution. The voting screen looks super cool. I love the 5-star voting approach. I just added some comments so the code can be a bit more mainteinable.

PR #241, opened 2023-05-16
Feature/Location Picker

pauromeu:
This PR will add a location picker for the creation of events and challenges. This PR hasn't been tested because the GoogleMap composable that has been used does not allow to do so. Also, note that in this PR a mini refactoring has been applied to the Event and Challenge class so that they allow to input the latitude and location when selecting it using the location picker.

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
+++ b/app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
@@ -27,7 +27,9 @@ fun LocationPicker(
     val cameraPositionState = rememberCameraPositionState {
         position = CameraPosition.fromLatLngZoom(EPFL, ZOOM_DEFAULT_VALUE)
     }
-    var markerState by remember { mutableStateOf<MarkerState?>(null) }
+    var markerState by remember {
+        mutableStateOf<MarkerState?>(null)
+    }
 
     Column {
         if (markerState == null) {


app/src/main/java/ch/epfl/sdp/cook4me/ui/challengedetailed/ChallengeDetailedScreen.kt
```diff
@@ -125,7 +125,7 @@ fun ChallengeDetailedScreen(
 
                 Spacer(Modifier.height(8.dp))
 
-                val position = LatLng(it.latLng.first, it.latLng.second)
```
nino-gerb, line 128: Nice this is easier to read

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
```diff
@@ -0,0 +1,60 @@
...
+        ) {
+            markerState?.let { state ->
+                Marker(state = state)
+            }
+        }
+    }
+}
+
+/*
+@Preview(showBackground = true)
```
nino-gerb, line 53: Maybe you should consider removing the comment if not used.

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
```diff
@@ -0,0 +1,60 @@
...
+import ch.epfl.sdp.cook4me.ui.map.Locations.EPFL
+import com.google.android.gms.maps.CameraUpdateFactory
+import com.google.android.gms.maps.model.CameraPosition
+import com.google.android.gms.maps.model.LatLng
+import com.google.maps.android.compose.GoogleMap
+import com.google.maps.android.compose.Marker
+import com.google.maps.android.compose.MarkerState
+import com.google.maps.android.compose.rememberCameraPositionState
+
+private const val ZOOM_DEFAULT_VALUE = 15f
```
nino-gerb, line 20: Looks good. You could add more comments to make it easier for readers to understand.

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+    var markerState by remember { mutableStateOf<MarkerState?>(null) }
+
+    Column {
+        if (markerState == null) {
+            Text("Click on the map to pick a location")
+        }
+        GoogleMap(
+            modifier = modifier,
+            cameraPositionState = cameraPositionState,
```
nino-gerb, line 38: For me its kinda hard to understand what it does exactly but as the onMapClick function looks ok to me. Maybe it would help the userexperience to have a confirmation of the location when it is picked

nino-gerb (approved):
Hello Paul. Congrats on you code. It looks as you as if you replaced the Point through Geolocation which provides more readablilty. The LocationPicker look clean. Maybe in the future the userexperience can be improved., with a confirmation on the pick or something like this. But it's not easy to write specific ciritic to the GoogleMaps function as they are really different from what I did so far. Keep up the good work.  Maybe you find some easy test to pump up the coverage, but do not waste time.

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
pauromeu: Great!

app/src/main/java/ch/epfl/sdp/cook4me/ui/map/LocationPicker.kt
pauromeu: That sounds good! I will create an issue to handle that in a future :)

PR #242, opened 2023-05-16
refactor challenge class

Tachi-67:
In the data class there are 2 maps:
- `participants: Map<String, Int>` to keep track of the participant's votes. e.g. `"haolong.li@epfl.ch" to 1`.
- `participantIsVoted: Map<String, Boolean>` to keep track of if the participant has voted. e.g. `"haolong.li@epfl.ch" to false`.

To update a challenge:
1. update the challenge locally. Call either `addParticipant`, `changeParticipantScore`, `changeParticipantIsVoted`. For example: `val updatedChallenge = addParticipant(challenge, "haolong.li@epfl.ch")`
2. submit the update to firestore. Call `challengeFormService.updateChallenge(id, updatedChallenge)` within a couroutine. Here the id is the id created by firestore upon creation of the challenge.

There are many ways one can query a challenge with the `getWithGivenField` function in the ChallengeFormService, it basically allows any query to the challenge collection given a field and the correcponding key. See comments in the code for more info, There are also other query methods implemented for the sake of convenience.

To get the name of the user, there is a `parseParticipantEmailToName` function. e.g.: `parseParticipantEmailToName("haolong.li@epfl.ch") = "haolong li"`


app/src/main/java/ch/epfl/sdp/cook4me/ui/challengeform/Challenge.kt
```diff
@@ -107,3 +122,31 @@ fun changeParticipantScore(challenge: Challenge, participant: String, scoreChang
...
+}
+
+/**
+ * Parse the email of a participant to a name
+ * @param email the email of the participant, e.g. could be retrieved from account service
+ * @return the name of the participant
+ * @sample parseParticipantEmailToName("darth.vader@epfl.ch") = "darth vader"
+ */
+fun parseParticipantEmailToName(email: String): String {
+    val name = email.split("@")[0]
```
nino-gerb, line 150: I think this is a good approach for a first time, but maybe we wanna add the profile name in an other sprint

app/src/main/java/ch/epfl/sdp/cook4me/ui/challengeform/Challenge.kt
```diff
@@ -107,3 +122,31 @@ fun changeParticipantScore(challenge: Challenge, participant: String, scoreChang
...
 }
+
+/**
+ * Change the is voted of a participant in the challenge.
+ * @param challenge the challenge to add the participant to
+ * @param participant the participant to add
+ * @param isvoted the updated isVoted status
+ * @sample updatedChallenge = changeParticipantScore(challenge, "participantName", true)
+ */
+fun changeParticipantIsVoted(challenge: Challenge, participant: String, isVoted: Boolean): Challenge {
```
nino-gerb, line 133: This function is nice. I was wondering if you have added a function to update the score.

app/src/main/java/ch/epfl/sdp/cook4me/application/ChallengeFormService.kt
```diff
@@ -25,6 +26,28 @@ class ChallengeFormService(
...
+     * Call either addParticipant, changeParticipantScore, changeParticipantIsVoted.
+     * For example: val updatedChallenge = addParticipant(challenge, "haolong.li@epfl.ch")
+     *
+     * 2. Submit the update to firestore.
+     * Call challengeFormService.updateChallenge(id, updatedChallenge) within a couroutine.
+     * Here the id is the id created by firestore upon creation of the challenge.
+     */
+    suspend fun updateChallenge(id: String, challenge: Challenge) {
+        try {
+            objectRepository.update(id, challenge)
```
nino-gerb, line 45: Nice that you used the objectRepository

nino-gerb (approved):
Overall your code looks nice. There are some function you could add like to update the score of a participant? But maybe it's not really needed. Thank you for the fast implementation! :-)

PR #246, opened 2023-05-17
Implement logic for swipe feature

bu-da:
This PR adds the swipe and matching logic for the tupperware feature. Furthermore, it replaces the hardcoded tupperware with the actual data. If a match happens a dialog pops up. We still need to implement that we open a chat in the background on a match, prettify the match dialog and the all done screen.

Note about the tests: I deleted the map tests, because they always ran into timeouts on my machine. Furthermore, the profile tests were flaky, so I commented them out. We definitely need to fix our tests next sprint. The commented out test case for the SwipeRepository also needs to be checked. Maybe storing data and checking it right after doesn't work all the time with firebase?


app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/TupperwareRepository.kt
```diff
@@ -1,43 +1,64 @@
...
+                user = it.user,
+                image = bytes
+            )
         }
     }
 
+    suspend fun getAllIdsByUser(email: String): Set<String> {
+        val result = store.collection(COLLECTION_PATH).whereEqualTo("user", email).get().await()
+        return result.map { it.id }.toSet()
+    }
```
dayan9265, line 51: Again maybe add doc or give more clean name to function, because when reading it might not be immediately obvious we are retrieving all tupperwares the user has created

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareform/CreateTupperwareScreen.kt
```diff
@@ -110,7 +110,7 @@ fun CreateTupperwareScreen(
             onCancel = onCancel,
             onSubmit = { title, description ->
                 scope.launch {
-                    repository.add(title, description, images)
+                    repository.add(title, description, images.first())
```
dayan9265, line 113: Does this mean only the first image is stored? if yes it would be a good idea to either force the user to give only one image or store all images the user proposes

app/src/androidTest/java/ch/epfl/sdp/cook4me/EmulatorUtils.kt
dayan9265, line 1: good idea to have added that since we repeat the same code in most tests 👍

app/src/androidTest/java/ch/epfl/sdp/cook4me/utils.kt
dayan9265, line 1: good idea to have these functions to makes tests more clear

dayan9265 (approved):
Good job, the code looks really good! Nice helper functions for the tests it makes testing more concise and readable 👍 
I realize that you added multiple `checkNotNull` but we never call theses functions with try / catch, we could look next sprint to fail gracefully if there are null values
And also, there seems to be quite a lot of flaky tests, I will add an issue to look at them and fix or remove them

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperwareform/CreateTupperwareScreen.kt
bu-da: Good catch, I actually already discussed this with John. We want to restrict the form so that only one image can be uploaded per tupperware (one image is better for the UX of the swipe screen. The functionality of stepping through multiple images would interfere with swipe behaviour, I already tried). This is just temporary, but I should have added a comment to clarify our intentions. Issue:  https://github.com/cook4me/android/issues/255

bu-da:
Thanks @dayan9265 for your code review. To respond to your comment about the checkNotNull calls. As far as I remember they only fail if the user is not logged in, so these actually should never happen (otherwise there is an implementation issue on our side). I don't really want to handle these errors gracefully, because this isn't the desired behaviour (in production we would get a notification / error, which then need to be fixed).

app/src/main/java/ch/epfl/sdp/cook4me/persistence/repository/ObjectRepository.kt
```diff
@@ -6,26 +6,27 @@ import com.google.firebase.firestore.FirebaseFirestore
...
 
-private const val MAX_RETRIEVED_DOCS = 20.toLong()
-
 open class ObjectRepository(
     private val store: FirebaseFirestore = FirebaseFirestore.getInstance(),
     private val objectPath: String = ""
 ) {
 
-    suspend fun <A : Any> addAndGetId(value: A): String {
+    suspend fun <A : Any> add(value: A): String {
```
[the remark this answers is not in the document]
bu-da: It's not uncommon to return a value (boolean or sth) on an add call, see e.g. kotlin collections: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/add.html. But I agree with you that a comment is helpful.

PR #253, opened 2023-05-18
Feature/ranking voting logic

nino-gerb:
I added a wrapper to the voting and the ranking. 
The ranking load a challenge from the database and displays the top users
The voting is a flow between voting screen and ranking it displays the voting screen if the user has not already voted.

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/cook4me/ui/challenge/VoteWrapper.kt
+++ b/app/src/main/java/ch/epfl/sdp/cook4me/ui/challenge/VoteWrapper.kt
@@ -101,7 +101,8 @@ fun VoteWrapper(
     when {
         isLoading -> LoadingScreen()
         isVoting -> challenge?.let {
-            VotingScreen(challenge = it,
+            VotingScreen(
+                challenge = it,
                 onVoteChanged = { challengeRet ->
                     challengeVote = challengeRet
                 },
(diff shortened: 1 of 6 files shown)


app/src/main/java/ch/epfl/sdp/cook4me/ui/challenge/VoteWrapper.kt
```diff
@@ -0,0 +1,137 @@
...
+
+@Suppress("ComplexMethod")
+@Composable
+fun VoteWrapper(
+    service: ChallengeFormService = ChallengeFormService(),
+    challengeId: String,
+    currentUser: String,
+    onBack: () -> Unit,
+) {
+    val coroutineScope = rememberCoroutineScope()
```
JohnTomasTaylor, line 26: It might be a good idea to refactor this code to make it more readable, there are too many indentation levels

app/src/main/java/ch/epfl/sdp/cook4me/ui/challenge/VoteWrapper.kt
```diff
@@ -0,0 +1,137 @@
...
+                // update challenge with new score
+                challengeVote?.let { challengeVote ->
+                    for (participant in challengeVote.participants) {
+                        challengeAfterVote = changeParticipantScore(
+                            challengeAfterVote!!,
+                            participant.key,
+                            participant.value
+                        )
+                    }
+                    challengeAfterVote = addParticipant(
```
JohnTomasTaylor, line 70: I think using !! has been banned according to company policies

JohnTomasTaylor (approved):
Hey nice work! However VoteWrapper probably needs to be refactored as it is not very readable.

PR #282, opened 2023-05-20
Refactor/chat

Tachi-67:
- I disabled the header (the channel list name, the pencil, the avatar) becuase they are kind of misleading.
- I enabled the action of long click on channel -> view channel info -> jump to the profile of the user chatting with.
- I keep record of user email in the chat user's extra data upon clicking inside of the chat feature.
- (Credits to @bu-da), Added the profile picture to the channel avatar
- Enable action of viewing chatmate's profile when clicking on header in message screen
- Fix bug: when trying to preview image from message screen, the app crashes. Now the preview is working fine.
- (not related to chat) add botton nav bar for the profile screen.
- Added an alert dialogue when too many connection attemps are made (this will lead to forever loading) and return to recipe feed screen.


app/src/main/java/ch/epfl/sdp/cook4me/ui/navigation/BottomNavigationBar.kt
```diff
@@ -115,7 +115,7 @@ val mainDestinations = listOf(
     BottomNavScreen.Recipes,
     BottomNavScreen.Events,
     BottomNavScreen.Challenges,
-    BottomNavScreen.Chat
+    BottomNavScreen.Chat,
```
bu-da, line 118: Nitpick: Comma could be removed

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/MessagesActivity.kt
```diff
@@ -33,22 +34,33 @@ class MessagesActivity : ComponentActivity() {
...
             ) {
                 MessagesScreen(
                     channelId = channelId,
                     messageLimit = 30,
-                    onBackPressed = { finish() }
+                    onBackPressed = { finish() },
+                    onHeaderActionClick = { channel ->
+                        val targetMember = channel.members.find { it.user.extraData["email"] != userEmail }
+                        val targetEmail = targetMember?.user?.extraData?.get("email")
+                        if (targetEmail != null) {
```
bu-da, line 47: Nitpick: You could the following instead:
```
targetEmail?.let {...}
```

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/MessagesActivity.kt
```diff
@@ -33,22 +34,33 @@ class MessagesActivity : ComponentActivity() {
...
+                ),
+                imageLoaderFactory = CoilImageLoaderFactory
             ) {
                 MessagesScreen(
                     channelId = channelId,
                     messageLimit = 30,
-                    onBackPressed = { finish() }
+                    onBackPressed = { finish() },
+                    onHeaderActionClick = { channel ->
+                        val targetMember = channel.members.find { it.user.extraData["email"] != userEmail }
```
bu-da, line 45: Nitpick: You're already using constants, which is good! The string `"email"` could also be one

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/ChannelScreen.kt
```diff
@@ -76,13 +80,18 @@ fun ChannelScreen(
...
+                        val intent = MessagesActivity.getIntent(context, channelId = channel.cid, userEmail = userEmail)
                         startActivity(context, intent, null)
                     },
                     onBackPressed = { onBackListener() },
-                    onHeaderAvatarClick = {
-                        client.disconnect(true).enqueue()
-                    },
+                    isShowingHeader = false,
+                    onViewChannelInfoAction = { channel ->
+                        val targetMember = channel.members.find { it.user.extraData["email"] != userEmail }
```
bu-da, line 87: Nitpick: getting the targetEmail seems to be same logic in this composable and in the MessagesActivity. Can you create a helper for it?

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/ChannelScreen.kt
```diff
@@ -29,25 +31,27 @@ fun ChannelScreen(
     ),
     accountService: AccountService = AccountService(),
     onBackListener: () -> Unit = {},
+    navController: NavController,
 ) {
-    // disconnecting the client before connecting again, otherwise will
-    // cause error: too many connections
-    runBlocking {
-        client.disconnect(true).enqueue()
-    }
```
bu-da, line 37: Question: We don't need this anymore?

bu-da (changes requested):
Thanks a lot for your work :-). I pulled your branch and there are a few things that needs to be addressed, before we're finally done with the chat:
* When I'm switching back and forth between the features, the chat is unsuable pretty fast. First it's shown that I'm disconnected in the Chat Header instead of the channel name. When I switch to another feature and back our loading screen is shown constantly. The chat feature is not really usable like this :-( How can we fix this?  
* If I use the search and type e.g. "2023" the channel "sdp2023cook4me" doesn't appear. Do you have control over this?
* If I click on the header in the chat I land on the profile, which is good. Didn't we say we want to see the navigation bar then? Or it would be good to have at least a back button or sth
* The loading screen of the chat is still shown. It doesn't look good if we have two loading screens.  I see two options:
      - overwrite somehow the `loadingContent` parameter of the [`ChannelList` composable](https://getstream.io/chat/docs/sdk/android/compose/channel-components/channel-list/#customization)
      - use their loading screen instead of ours in the whole app
* There aren't any tests
* Question: Can we apply our colors / design to the screens?
Other issue: I still don't like that we have multiple Activities now, but as long as it is functional this is not a priority.

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/ChannelScreen.kt
Tachi-67: no, I tried to move the disconnect into a queue (like you suggested) and it turned out this call never succeeds

app/src/main/java/ch/epfl/sdp/cook4me/Cook4MeApp.kt
```diff
@@ -60,7 +60,7 @@ fun Cook4MeApp(
 
     val isOnline = networkCallback.isOnline.value
 
-    val screensWithBottomBar = mainDestinations.map { it.route }
+    val screensWithBottomBar = mainDestinations.map { it.route } + Screen.ProfileScreen.name
```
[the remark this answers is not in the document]
Tachi-67: This is to show button nav bar also in the user's own profile screen.

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/ImageLoaderDecorator.kt
```diff
@@ -0,0 +1,54 @@
...
+import coil.ComponentRegistry
+import coil.ImageLoader
+import coil.disk.DiskCache
+import coil.memory.MemoryCache
+import coil.request.DefaultRequestOptions
+import coil.request.Disposable
+import coil.request.ImageRequest
+import coil.request.ImageResult
+
+const val PROFILE_IMAGE_PREFIX = "load profile image"
```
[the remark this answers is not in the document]
Tachi-67: done

app/src/main/java/ch/epfl/sdp/cook4me/ui/navigation/Cook4meNavHost.kt
```diff
@@ -125,6 +126,20 @@ fun Cook4MeNavHost(
             )
         }
         composable(route = Screen.ProfileScreen.name) { ProfileScreen() }
+        // changing the profile screen to take an id as an argument(email)
+        // if nothing is passed, it will show the current user's profile
+        composable(
+            route = "${Screen.ProfileScreen.name}/{userId}",
+            arguments = listOf(navArgument("userId") { type = NavType.StringType })
+        ) { backStackEntry ->
+            val userId = backStackEntry.arguments?.getString("userId")
```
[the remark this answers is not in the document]
Tachi-67: done

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/MessagesActivity.kt
Tachi-67: done

app/src/main/java/ch/epfl/sdp/cook4me/ui/chat/MessagesActivity.kt
Tachi-67: might be a good idea to leave indexes as clear text for readability

bu-da (approved):
I still don't like that we can't avoid the connection timeouts, but we at least inform the users about it. Thanks for adding the dialog and clarifying my questions in your comment and today at the meeting

PR #283, opened 2023-05-20
sketch a draft how we can profile images

bu-da:
This PR sketches a solution how we can load the profile image for the chat feature. Unfortunately, we can't store them on the getstream server, so we need to implement a workaround. There is also a potential race condition I tried to fix in the ChannelScreen (`client.disconnect(...)` should probably always be executed / terminated before `client.connectUser(...)`)


Tachi-67 (commented):
Thank you for the code!!! Really help a lot!

PR #286, opened 2023-05-21
extend firebase functions to be able to use custom mapper

bu-da:
I adapted the firebase utility functions to be able to use custom mappers. @Tachi-67 The mapping should actually happen in the repositories and not in the data classes. Unfortunately, I didn't have time to refactor this.


Tachi-67 (approved):
I dont get what is not working because the branch is named after fix, but yeah it looks good.

PR #290, opened 2023-05-21
add back common form tests and reorder them

bu-da:
This PR adds the common form tests back. I moved them all to the same package.


Tachi-67 (approved):
lgtm

PR #293, opened 2023-05-22
Fix/user tests

bu-da:
This PR reorders the package structure to group all related user ui components together (affects login, signup and profile). Additionally, the tests regarding these screens are commented in again and improved.


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/user/signup/AddProfileInfoScreenTest.kt
```diff
@@ -0,0 +1,214 @@
...
+        assert(profileViewModel.profile.value.favoriteDish == FAVFOOD_INPUT)
+        assert(profileViewModel.profile.value.allergies == ALLERGIES_INPUT)
+        assert(profileViewModel.profile.value.bio == BIO_INPUT)
+
+        runBlocking {
+            // check that the user is created
+            auth.signInWithEmailAndPassword(signUpViewModel.profile.value.email, PASSWORD_INPUT)
+                .await()
+            assert(auth.currentUser != null)
+        }
```
dayan9265, line 152: I know this test was already existing, but I think we should try to avoid the runBlocking in tests and also detail but the test is quite big testing different parts of the code and it has a simple name with not a lot of meaning `testTextFieldsInput`

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/user/signup/SignUpViewModelTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        composeTestRule.waitUntil(timeoutMillis = 5000) {
+            !profileViewModel.isLoading.value
+        }
+
+        // check that the user is created correctly
+        assert(profileViewModel.profile.value.name == USERNAME)
+        assert(profileViewModel.profile.value.favoriteDish == FAVORITE_DISH)
+        assert(profileViewModel.profile.value.allergies == ALLERGIES)
+        assert(profileViewModel.profile.value.bio == BIO)
+        assert(profileViewModel.profile.value.email == EMAIL)
```
dayan9265, line 131: also detail, since it comes form ancient test, but using assertThat() would give us more information in case the test doesn't pass

dayan9265 (approved):
Looks good to me, thanks for making the tests working again :)

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/user/signup/SignUpViewModelTest.kt
bu-da: Very good point, I'll adapt it

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/user/signup/AddProfileInfoScreenTest.kt
bu-da: I renamed it to `completeSignUpFlowTest`. Unfortunately, I don't have a better approach to avoid using `runBlocking` in this test (at least it's not wrapped in a `runTest` or `runBlocking` call).

PR #294, opened 2023-05-22
Reorder package structure for recipes and add tests back

bu-da:
This PR reorders the recipe files and adds its tests back.


nino-gerb (approved):
Looks like you refactored a lot of the tests. It's make sense to put commonly used packages in the common package. It seems that there are still a lot of tests missing but we're making a lot of progress putting them pack. Keep up the good work!

PR #295, opened 2023-05-22
Add back the navigation tests

bu-da:
This PR adds back the navigation tests


app/src/androidTest/java/ch/epfl/sdp/cook4me/TestPermissionStatusProvider.kt
dayan9265, line 1: not sure to understand this file, since I see no tests but only composable as function, maybe add some documentation to this class to make more clear what is the goal

dayan9265 (approved):
Except one file where I am not sure what is intended, LGTM 👍

app/src/androidTest/java/ch/epfl/sdp/cook4me/TestPermissionStatusProvider.kt
bu-da: Thanks for your review. This was done by Pau, not by me. Unfortunately, I don't have time to improve its documentation or readability.

PR #296, opened 2023-05-22
Reorder tupperware files and add tests back

bu-da:
This PR reorders the tupperware files and comments the test back in.


nino-gerb (approved):
It looks like you made just a reordering and you put the tests back. I did not a detailed pr as it was already made previously. But it's nice to see that our coverage is back up where it belongs and that you put the hard work in to correct our complete test suite.

PR #297, opened 2023-05-22
Fix/map tests

bu-da:


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/map/MapScreenTest.kt
```diff
@@ -1,134 +1,119 @@
...
-//    @Test
-//    fun testNoEventSelectedWhenStartingScreen() {
-//        initMap()
-//        // To be shown when no events are displayed
-//        composeTestRule.onNodeWithText("Select an event").assertIsDisplayed()
-//    }
+    @After
+    fun cleanUp() {
+        runBlocking {
+            eventRepository.deleteAll()
```
nino-gerb, line 67: I encountered once a problem running my test suit locally where it did not connect correctly to the emulator and then connect to the real database. This could create an issue? Maybe it would be more secure to just delete the added events not every...

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/map/MapPermissionWrapper.kt
```diff
@@ -1,46 +1,43 @@
...
+//        composeTestRule.setContent {
+//            MapPermissionWrapper(
+//                permissionStatusProvider = permissionStatusProvider,
+//                modifier = Modifier.fillMaxSize()
+//            )
+//        }
+//
+//        composeTestRule.onNodeWithText("EPFL").assertIsDisplayed() // Map is shown
+//    }
+//
```
nino-gerb, line 29: Still not working?

nino-gerb (approved):
As always. Good improvement. I added a comment about an issue i had once with the emulator and a parts of the code that could be problematic with this way of using the emulator. But I think you resolved this problem. Maybe think about linking it to an issue.

PR #298, opened 2023-05-22
Fix permissions test

dayan9265:


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/map/MapPermissionWrapper.kt
```diff
@@ -1,43 +1,49 @@
...
+
+    @Test
+    fun assertMapIsDisplayedWhenLocationIsEnabled() {
+        val permissionStatusProvider = TestPermissionStatusProvider(mapOf("location" to Pair(true, false)))
+
+        composeTestRule.setContent {
+            MapPermissionWrapper(
+                permissionStatusProvider = permissionStatusProvider,
+                modifier = Modifier.fillMaxSize(),
+                isOnline = true,
```
pauromeu, line 28: Question: Why do you set isOnline to true? This will try to display the user location in the test and may crash it.

pauromeu (approved):
LGTM :) Good job, I just added a little comment

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/map/MapPermissionWrapper.kt
```diff
@@ -1,43 +1,49 @@
...
+            )
+        }
+
+        composeTestRule.onNodeWithText("EPFL").assertIsDisplayed() // Map is shown
```
pauromeu, line 32: Remark: In a future, we will remove the button with text "EPFL" on it so this test will crash. Maybe another method should be found to assert that the map is displayed. We can think of it when refactoring the map screen.

pauromeu (approved):
Very good job :) LGTM. I just added a small comment.

app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/map/MapPermissionWrapper.kt
dayan9265: Yes true, when refactoring we should be careful to modify the tests as well

PR #299, opened 2023-05-23
include logo

bu-da:
This PR includes our new logo as app icon, launcher icon etc. We may slightly adapt it the following the days, but with this PR we get at least a feeling how it will look like at the end.


dayan9265 (approved):
Feels nice to finally have our own logo, looks good to me :)

PR #302, opened 2023-05-23
add placeholder and loading screens

bu-da:
This PR adds the placeholder screens for the tupperware and recipe feature. The placeholder screen is fully prepared for the challenge feature as well, but it needs to done after the merge of https://github.com/cook4me/android/pull/254. Additionally, the PR shows the existing loading screen in the recipe and profile feature. It also needs to be added to the challenge feature.


app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperware/swipe/CircleButton.kt
```diff
@@ -18,20 +17,19 @@ import androidx.compose.ui.unit.dp
...
             .size(56.dp)
-            .border(2.dp, if (enabled) MaterialTheme.colors.primary else Color.Gray, CircleShape),
+            .border(2.dp, color, CircleShape),
         onClick = onClick,
-        enabled = enabled
     ) {
         Icon(
             icon, null,
-            tint = MaterialTheme.colors.onPrimary
+            tint = Color.White
```
dayan9265, line 32: why remove the ability to have a non clickable button, in my opinion it can be useful? (when offline or when all matches has been done?)

app/src/main/java/ch/epfl/sdp/cook4me/ui/recipe/feed/BottomBar.kt
```diff
@@ -0,0 +1,78 @@
...
+ * it takes a boolean as parameter, true if top recipes button is clicked, false otherwise
+ */
+@Preview(showBackground = true)
+@Composable
+fun BottomBar(onButtonClicked: (Boolean) -> Unit = {}) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(50.dp)
+            .background(Color.White),
```
dayan9265, line 37: Don't if we want it is good practice to have all colors used being from our resource palette or if it is okay to have also hard-coded ones

app/src/main/java/ch/epfl/sdp/cook4me/ui/common/PlaceholderScreen.kt
```diff
@@ -0,0 +1,36 @@
...
+    @StringRes text: Int
+) {
+    Column(
+        modifier = Modifier.fillMaxSize(),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Image(
+            painter = painterResource(image),
+            contentDescription = "${stringResource(text)} placeholder screen",
```
dayan9265, line 30: " placeholder screen" could be put in the string resource

dayan9265 (approved):
Nice refactoring of parts of our code, globally looks really good 👍

app/src/main/java/ch/epfl/sdp/cook4me/ui/tupperware/swipe/CircleButton.kt
bu-da: When all matches are done, where not showing the buttons at all. If we're offline, we could show the placeholder screen?

app/src/main/java/ch/epfl/sdp/cook4me/ui/recipe/feed/BottomBar.kt
bu-da: Good point, actually we should use the colors from our color palette and the buttons don't really conform to our general style. As John has already changed the implementation on another branch we don't need to change it.

PR #305, opened 2023-05-23
use loadingScreen for profiles

Tachi-67:
I changed the weird blue dot loading indicator in the profile screen and used our standard loading screen.


nino-gerb (approved):
LGTM

PR #308, opened 2023-05-23
remove map permission wrapper tests, which cause issues

bu-da:
This PR removes the flaky MapPermissionWrapper tests. We don't have a solution for them at the moment, so we get rid of it.


Tachi-67 (approved):
lgtm

PR #310, opened 2023-05-24
Refactor/loading screen

Tachi-67:
Making the loading screen more beautiful. I used the launcher icon (the chef pic) and gave some simple image effects.

Looks much better than the previous old school stuff


JohnTomasTaylor (approved):
Nice thank you for making the world a little more beautiful

PR #314, opened 2023-05-24
Refactor/chat tests

Tachi-67:
added some tests just for the coverage


JohnTomasTaylor (approved):
Hey, thanks for the work, new tests are always welcome. However If you have the time I think you should consider rewriting them to be a little more meaningful. Right now they don't seem to test much. Correct me if I'm wrong of course

PR #317, opened 2023-05-25
Add sign up flow back

bu-da:
This PR adds back the sign up flow and applies our new styling to it. Unfortunately, I couldn't reuse the tests. I'll write new ones and will add them via a new PR


app/src/main/java/ch/epfl/sdp/cook4me/persistence/model/Profile.kt
```diff
@@ -3,7 +3,6 @@ package ch.epfl.sdp.cook4me.persistence.model
 @Suppress("DataClassShouldBeImmutable")
 data class Profile(
     var email: String = "",
-    var name: String = "",
```
Tachi-67, line 6: we should have names in the profile.

Tachi-67 (changes requested):
lgtm, but we need to have the name in the profile.

bu-da:
The new tests are added via this PR: https://github.com/cook4me/android/pull/334

PR #319, opened 2023-05-25
Fix/offline mode

dayan9265:


app/src/androidTest/java/ch/epfl/sdp/cook4me/ui/user/profile/ProfileScreenTest.kt
```diff
@@ -86,24 +87,4 @@ class ProfileScreenTest {
 
         composeTestRule.onNodeWithTag("defaultProfileImage").assertExists()
     }
-
```
JohnTomasTaylor, line 89: It's okay to do these kinds of fixes now when we are in a rush but don't forget to add the fix as an issue to the backlog

app/src/main/java/ch/epfl/sdp/cook4me/ui/user/profile/ProfileScreen.kt
```diff
@@ -15,30 +15,58 @@ import androidx.compose.foundation.shape.CircleShape
...
 ) {
-    val profile = profileViewModel.profile.value
-    val userNameState = rememberSaveable { mutableStateOf("") }
-    val isLoading = profileViewModel.isLoading.value
-    val image = profileViewModel.profileImage.value
+    var profile by remember { mutableStateOf(Profile()) }
+    var isLoading by remember { mutableStateOf(true) }
+    var image by remember { mutableStateOf(EMPTY) }
+
+    LaunchedEffect(Unit) {
```
JohnTomasTaylor, line 51: Good refactoring but try to avoid using suppress as it can lead to bad code

PR #322, opened 2023-05-25
UI/beautify permission manager

pauromeu:


nino-gerb (approved):
LGBT but you could add string to stringresource and remove preview if you have time

PR #324, opened 2023-05-25
Refactor/event form

JohnTomasTaylor:
This PR deletes location and privacy fields from the event class, fixes navigation when creating a new event, and deletes the hardcoded user in the voteWrapper


Tachi-67:
OHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHH

PR #335, opened 2023-05-27
renew chat api key in .cirrus.yml to let CI know

Tachi-67:
I renewed the chat api key. Each api key needs to be updated every 30 days if we don't pay to the service.

This pr is to inform the CI of the updated key, it has nothing to do with the app's features.


nino-gerb (approved):
<3

